PR #30, opened 2018-07-07
Some protobuf housekeeping

bocon13:
- Minor fixes to `compile_protos.sh`
- Adding explicit `go_package` and making proto imports relative
This is important because for non-Go languages, users will likely not have the fully qualified path including `github.com/openconfig/gnmi` as part of their local workspace.


bocon13:
@robshakir Does this break anything?

robshakir:
@bocon13 Thanks for the changes here. We'll review them. Currently, this repo is copied from a primary source, so we'll likely make the required changes there and copy them into this repo too.

Thanks,
r.

PR #49, opened 2019-02-13
Add extra CLI auth args and README doc

remingtonc:
Adds some ability for those running `gnmi_cli` to use either a JSON file containing `Username`/`Password` or directly as args with `-insecure_username`/`-insecure_password` for authentication.

Also adds some usage documentation to make reference tooling easier to onboard/use. The docs should probably be verified. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,155 @@ To install the CLI, run
 
     go get -u github.com/openconfig/gnmi/cmd/gnmi_cli
 
+### CLI Usage
+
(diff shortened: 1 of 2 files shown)


remingtonc:
Closing and re-forking to tidy up.

remingtonc:
> In example.cfg the key should be "addresses" instead of "address"

@manojgn Thank you for catching that, fixed in separate fork/pull #59

PR #55, opened 2019-05-12
corrected typos in compile_protos.sh

Linerd:
The original file compiles gnmi_ext twice.


robshakir:

robshakir:
Thanks for this change, I've merged the change upstream.

PR #57, opened 2019-07-29
Update example to match the updated Target proto

philodavies:
Example now uses `addresses` instead of `address`. This reflects the change
made to target.proto done by @mkhsueh


philodavies:
Can this be merged? The example is nice to have, but the change to the protobuf has caused it to error without this update.

cc @mkhsueh @robshakir @gcsl

robshakir:

robshakir:
Thanks for this - I've now merged this.

PR #66, opened 2020-01-15
connection: fix goroutine in TestConcurrentDialErr()

alrs:
`TestConcurrentDialErr()` was using `testing.T.Fatal()` inside a goroutine. `T.Fatal()` is a convenience function that combines `T.Log()` and `T.FailNow()`. From the go doc for `T.FailNow()`:

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

This moves `t.Fatal()` out of the goroutine, and puts the check for a `nil` error together with the check for equality between the two errors sent down the `errs` channel. It also logs the found errors.

### DIFF
--- a/connection/connection_test.go
+++ b/connection/connection_test.go
@@ -23,8 +23,8 @@ import (
 	"testing"
 	"time"
 
-	"google.golang.org/grpc"
 	"github.com/openconfig/gnmi/unimplemented"
+	"google.golang.org/grpc"
 
 	pb "github.com/openconfig/gnmi/proto/gnmi"
 )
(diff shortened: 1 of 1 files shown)


mkhsueh:
Hi @alrs, thanks for your pull request.

Due to testing/compatibility limitations, changes must currently be authored within Google and pushed out to GitHub.

We will mirror this change internally, and credit your PR in the commit description if approved.

mkhsueh:
See https://github.com/openconfig/gnmi/commit/c09d36ff6a4ac12cdbaf95b0e675001f264ffe36 for reference.

PR #70, opened 2020-02-20
Fixed proto path in compile_protos.sh

0xf1d0:


pixsll:
@googlebot I signed it!

PR #71, opened 2020-03-23
Update protobufs to use new protobuf library.

robshakir:
```
 * (M) {proto/*,testing/fake/proto/*}
  - Update protobufs compiling using new API.
 * (M) {go.mod,go.sum}
  - Update Go modules.
```


robshakir:
Thanks @mkhsueh !

r.

PR #76, opened 2020-05-27
Added extension ID to be used for exporting Juniper telemetry header.

amanand:
The extension ID will be used to export the telemetry header used by Juniper for exporting key information for debugging as well as for use by collectors. Details about the message that is exported is present here:
https://github.com/Juniper/telemetry/blob/master/20.1/20.1R1/protos/GnmiJuniperTelemetryHeader.proto

### DIFF
--- a/proto/gnmi_ext/gnmi_ext.proto
+++ b/proto/gnmi_ext/gnmi_ext.proto
@@ -47,6 +47,11 @@ enum ExtensionID {
   // MUST link to a reference describing their implementation.
 
   // Juniper Telemetry header
+  // The extension proto file is named as GnmiJuniperTelemetryHeaderExtension.proto
+  // for all Junos releases starting from 21.3
+  // and it is found in each release folders of the following github repo
+  // https://github.com/Juniper/telemetry
+  // Implementation details can be found at https://kb.juniper.net/KB36506
   EID_JUNIPER_TELEMETRY_HEADER = 1;
 
   // An experimental extension that may be used during prototyping of a new
--- a/proto/gnmi_ext/gnmi_ext.proto
+++ b/proto/gnmi_ext/gnmi_ext.proto
@@ -46,6 +46,9 @@ enum ExtensionID {
   // New extensions are to be defined within this enumeration - their definition
   // MUST link to a reference describing their implementation.
 
+  // Juniper Telemetry header
+  EID_JUNIPER_TELEMETRY_HEADER = 1;
+
   // An experimental extension that may be used during prototyping of a new
   // extension.
   EID_EXPERIMENTAL = 999;


amanand:
@googlebot I signed it


Thanks,
Amish

proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -46,6 +46,9 @@ enum ExtensionID {
   // New extensions are to be defined within this enumeration - their definition
   // MUST link to a reference describing their implementation.
 
+  // Juniper Telemetry header
+  EID_JUNIPER_TELEMETRY_HEADER = 1;
```
robshakir, line 50: Can you please add a URL to the proto definition / implementation of this header as per the above comment?

amanand:
Hi Rob,

The timestamps in the extension proto are informational and useful in debugging any issues related to latency or KPI evaluation of the telemetry data.
The event_timestamp indicates the time when the  event occurs on the device whereas the Notification timestamp is when the device emits the data to the collector.
The export_timestamp indicates the time when complete telemetry data corresponding to either event or periodic data is sent from the producer.
Difference between export_timestamp and event_timestamp will indicate the total time taken to fire the event at producer.
Difference between export_timestamp and the timestamp in Notification will indicate the internal latency in the infra.

Thanks,
Amish

amanand:
Hi @robshakir 
Can you please share any updates on this?
Thanks
Amish

gcsl:
On Fri, Jun 19, 2020 at 3:16 PM Amish Anand <notifications@github.com>
wrote:

> Hi Rob,
>
> The timestamps in the extension proto are informational and useful in
> debugging any issues related to latency or KPI evaluation of the telemetry
> data.
> The event_timestamp indicates the time when the event occurs on the device
> whereas the Notification timestamp is when the device emits the data to the
> collector.


I believe this is the reason Rob was seeking clarification.  The timestamp
in the notification is supposed to be the event timestamp, tightly coupled
to the data (counter or event), not a serialization or sending timestamp.
This is one of the primary motivations for moving to Streaming Telemetry
over SNMP in that accurate rates for counters cannot be computed or latency
derived if there is no known timestamp associated with the data or event.
This is especially problematic when there can be various levels of caching
between the event and the serialization of the data.  If the timestamps
derived for gNMI Notifications are not event timestamps in the existing
implementation, that should be fixed.  Serialization timestamps might be
useful for debugging and could be added as an extension to help identify
certain types of problems.

robshakir:
Yes - it sounds like `event_timestamp` is what should be in the `Notification` of the gNMI message - and that this proto should carry `export_timestamp` and `gnmi_export_timestamp` or something similar for the serialisation tracing/debugging.

If we merge with the current behaviour, then this essentially means that we're requiring vendor-specific implementation in this interface, since `event_timestamp` should be what is in the `Notification`.

For the record, I think this serialisation debugging/tracing is potentially a useful extension (and is something that we've discussed internally), such that I think there's room for a well-known extension here.

amanand:
The difference between the Notification timestamp and event_timestamp will be noticeable only in case of Routing sensors due to queuing latency for those sensors. However, I agree the Notification timestamp should indeed be the event timestamp and the one in extension should be termed differently. I will update the proto accordingly and update Juniper implementation accordingly.
I agree that this could be well-known extension with possible couple of fields omitted out. Please do share any timelines for publishing the extension file. We will be keen to take part in any discussions around it.

robshakir:
Thanks for updating the proto -- please ping the PR when the change is made.

I'm happy for you to start a well-known extension issue to discuss which fields we might want to add here -- alternatively, I can start the issue. Once it's open (with a top-level proposal) I can ping other interested and potentially interested folks.

amanand:
Hi Rob,
Sorry for pretty late action on this one. I have updated the pull request with the requested details. 

Thanks
Amish

robshakir:
@amanand Is this still live? Apologies, I was on medical leave when you updated this, so I didn't see the change.

PR #84, opened 2020-11-11
grpc.WithTimeout deprecated

mehrdadrad:
As grpc.WithTimeout deprecated, I just removed it and added context.WithTimeout instead.
https://godoc.org/google.golang.org/grpc#WithTimeout

### DIFF
--- a/client/gnmi/client.go
+++ b/client/gnmi/client.go
@@ -30,14 +30,14 @@ import (
 	"time"
 
 	log "github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


robshakir:
This looks reasonable to me, I will propose this internally and make subsequent updates to this PR with any comments.

robshakir:

robshakir:
Thanks for this change, I've merged it upstream in d755837

PR #90, opened 2021-02-04
embed tunnel to gnmi collector, fake client and add a latency package

jxx-gg:
- Add a latency package to support latency stats for a given set of time windows. Allow to register IntValue (int64) type of metadata.
- Add a tunnel package as a wrapper for grpc tunnel
- embed the grpc tunnel to gnmi collector and fake client


jxx-gg:
adding @googlebot to rescan CLA

jxx-gg:
adding @googlebot to rescan CLA

PR #101, opened 2021-05-11
Add/change java options to proto sources

norrisjeremy:
Add/change java options to proto sources:
- Update `java_package` option in `gnmi.proto` to better align with `go_package` option.
- Add various java options to remaining proto sources.


norrisjeremy:
@googlebot I signed it!

PR #103, opened 2021-05-24
Fix test

gescheit:
Fix tests and add running if then in github action


gescheit:
@robshakir ping

gescheit:
@marcushines-zz

Issue #45, opened 2018-12-20
How to set DSCP value on gRPC stream in C++?


infn-ke:
The gNMI standard has a `qos` field that is used to mark the DSCP value of the gRPC stream. Does gRPC C++ API support setting DSCP on the stream? Doing it with `setsockopt` would require us to get the socket file descriptor...

robshakir:
Based on implementation discussions, we're currently in the process of working on an update that deprecates the `qos` field in the proto - based on gRPC support. We'll rather prefer a statically configured DSCP for the whole service. Operational experience has shown that there's little benefit of per subscription QoS.

infn-ke:
So an implementation has to remark the frames from gRPC externally? Or gRPC will add support for service based DSCP?

PR #116, opened 2021-10-26
Marking of 'QOSMarking' for deprecation

earies:
Per https://github.com/openconfig/gnmi/issues/45, I would like to initiate
deprecation of the `QOSMarking` field in favor of static configuration per each
service or grpc-server as a whole.

YANG configuration node support will need to still be discussed since services
are currently a leaf-list w/o per service configuration attributes.


ashu-ciena:
Is this change going forward ?

Issue #124, opened 2022-05-20
import error of gnmi_collector.go


peichengATgit:
There exist import errors for the below lines 
	"github.com/openconfig/gnmi/manager/manager"
	"github.com/openconfig/gnmi/tunnel/dialer"

gnmi_collector.go:40:2: no required module provides package github.com/openconfig/gnmi/manager/manager; to add it:
        go get github.com/openconfig/gnmi/manager/manager
gnmi_collector.go:43:2: no required module provides package github.com/openconfig/gnmi/tunnel/dialer; to add it:
        go get github.com/openconfig/gnmi/tunnel/dialer

It should be
	"github.com/openconfig/gnmi/manager"
	"github.com/openconfig/gnmi/tunnel"

There was similar fix previously, but these two offending lines were added after it.
https://github.com/openconfig/gnmi/commit/f4fab1ef8dc4aeb55e6df99debc1d5e313b0b33f

robshakir:
Thanks for pointing this out -- I've staged a fix internally for this.

PR #125, opened 2022-05-24
Add Go CI.

robshakir:
It looks like the go workflow was removed at some point, potentially due to an upstream sync issue.

Currently, this CI fails because of a fix being required for #124 -- see upstream fixes.


marcushines (dismissed):

marcushines:
static analysis is failing?

PR #126, opened 2022-07-07
Deprecation of alias support for gNMI 0.9.0

earies:
- Deprecation of alias per https://github.com/openconfig/reference/pull/148
- Regeneration of stubs

`Alias` support in gNMI was removed from the specification however the `.proto` was never updated to reflect these changes.  This PR addresses removal (deprecation) of alias support in the proto IDL for client/server implementations.

References:
* https://github.com/openconfig/reference/pull/148


hellt:
this is a considerable change which backfired at us since the published go package to this date is unversioned, which makes it quite hard to pin dependencies.

Currently, I see that the generated stubs are wiped from aliases entirely (not deprecated, but moved to reserved). https://pkg.go.dev/github.com/openconfig/gnmi@v0.0.0-20220920173703-480bf53a74d2/proto/gnmi#SubscriptionList

I think we really need to start tagging this repo with semver tags matching the spec version, otherwise downstream users are doomed to have issues with pinning dependencies and understanding which generated collection corresponds to which gnmi version.

hellt:
@marcushines @robshakir this commit https://github.com/openconfig/gnmi/commit/480bf53a74d21bb0a82d5d716264874de1070120#diff-49fc5824be3cfb8d64321cc64142f0568a3039a55da00d2aaed8fd766369854c changed gnmi.proto and generated stubs in a non-backwards compatible way without bumping the service version.

Currently, the statement outlined in the  https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#grpc-network-management-interface-gnmi header renders itself false, as the service proto and stubs do not conform to the 0.8.x version

hellt:
@earies I see that @marcushines just released 0.9.0, and one thing that is quite different from this proposal is that `alias` and `use_aliases` were not deprecated, but moved to reserved fields of the protos.
This effectively makes those fields absent from the generated stubs entirely...

PR #127, opened 2022-07-08
applied gofmt

hellt:
to fix failing static analysis https://github.com/openconfig/gnmi/pull/125#issuecomment-1136303307

/cc @marcushines 

PS. done with `gofmt ./...` from the root of the repo


hellt:
superseded by #128

PR #128, opened 2022-07-08
applied gofmt

hellt:
to fix failing static analysis https://github.com/openconfig/gnmi/pull/125#issuecomment-1136303307

/cc @marcushines 

PS. done with `gofmt ./...` from the root of the repo

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -27,14 +27,14 @@ import (
 	"time"
 
 	log "github.com/golang/glog"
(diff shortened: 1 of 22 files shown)


marcushines:
This unfortunatlely is all autogenerated off of an internal repo like you thought - i will look into seeing if there is anything we can do to make the output better - filing an internal FR to see if we can correct this
this can't be done in the current PR though

PR #135, opened 2023-02-16
Fix proto generation script and improper imports.

asauravpica8:
The `complie_protos.sh` script seems to be failing right now because of the incorrect import paths in the `protoc` command. Some imports in the `gnmi.proto`, `target.proto` and `fake.proto` also fail.

There are other instances of these failures being talked about in [blogs like these](https://karneliuk.com/2020/05/gnmi-part-3-using-grpc-to-collect-data-in-openconfig-yang-from-arista-eos-and-nokia-sr-os/).


asauravpica8:
@robshakir PTAL! 🙏

asauravpica8:
While we figure out the container-based proto generations as the right architecture going forward, could we accept this PR to fix broken builds in the meantime? @robshakir @n0shut @marcushines Appreciate your time 🙏

Issue #134, opened 2023-02-15
Proto includes failing and compile_protos.sh script errors out.


asauravpica8:
Checking out the main branch and running `./compile_protos.sh` fails with the following error:

```
/Users/amitsaurav/Work/go/src/github.com/google/protobuf/src: warning: directory does not exist.
/Users/amitsaurav/Work/go/src/github.com/googleapis/googleapis: warning: directory does not exist.
github.com/openconfig/gnmi/proto/gnmi/gnmi.proto: File not found.
testing/fake/proto/fake.proto:23:1: Import "github.com/openconfig/gnmi/proto/gnmi/gnmi.proto" was not found or had errors.
testing/fake/proto/fake.proto:106:12: "gnmi.SubscribeResponse" is not defined.
```

I have a change in my local branch but unable to submit a PR. Here are the changes that fix the compile_protos.sh script and go.mod file:

```
diff --git a/compile_protos.sh b/compile_protos.sh
index e0b5116..f0466f4 100755
--- a/compile_protos.sh
+++ b/compile_protos.sh
@@ -16,13 +16,16 @@

 set -euo pipefail

+# this function will get the location of a fully-qualified go import where it is stored in the local, module cache
+moddir() { go list -mod="" -m -f '{{ .Dir }}' "$1"; }
+
 # Go
 if ! which protoc-gen-go-grpc; then
   go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 fi
-protobufsrc=${GOPATH}/src/github.com/google/protobuf/src
-googleapis=${GOPATH}/src/github.com/googleapis/googleapis
-proto_imports_go=".:${protobufsrc}:${googleapis}:${GOPATH}/src"
+protobufsrc=$(moddir github.com/google/protobuf)
+googleapis=$(moddir github.com/googleapis/googleapis)
+proto_imports_go=".:${protobufsrc}:${googleapis}:$(pwd)/proto/gnmi/"
 protoc -I=$proto_imports_go --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative,require_unimplemented_servers=false testing/fake/proto/fake.proto
 protoc -I=$proto_imports_go --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative,require_unimplemented_servers=false proto/gnmi/gnmi.proto
[...]

PR #136, opened 2023-02-18
Container-based proto compilation

n0shut:
Hi @robshakir @marcushines 

I would like to propose a hermetic way to compile protos used in this repository which is based on the container approach.

Currently, the compile-protos.sh script has no pinned dependencies resulting in users defaulting to whatever latest/oldest version they have installed on their machines when updating the protos.

Not only it adds unnecessary churn to the PRs, it also makes the process overly complicated. The dev machine should have protos installed in the specified directories and we see that people struggle with that - #134.

In the newly added compile-protos-docker.sh I introduced a container-based workflow where I leverage ghcr.io/srl-labs/protoc:0.0.3 that has pinned dependencies for protoc and the relevant grpc plugins - https://github.com/srl-labs/protoc-container/blob/main/Dockerfile

Using this container every user of gnmi repo will be guaranteed to compile the protos in one and only possible way, without the need to keep local dependencies and maintain their lifecycle.

At the time of this writing the latest version of protoc and plugins were used. I hope you will find that useful.


compile-protos-docker.sh
```diff
@@ -0,0 +1,61 @@
+# This script is used to generate Go and Python language bindings for the proto files in this repository.
+# It uses srl-labs/protoc:0.0.3 docker image to generate the bindings. The image contains the pinned versions of the
+# protoc compiler and the Go and Python GRPC plugins.
+
+DOCKER_CMD="docker run -v $(pwd):/in \
+  -v $(pwd)/proto:/in/github.com/openconfig/gnmi/proto \
+  ghcr.io/srl-labs/protoc:0.0.3"
```
robshakir, line 7: For my edification, how does this map to the `protoc` version that is in the container?

robshakir (commented):
This looks great, thanks @n0shut. One quick question in the review.

compile-protos-docker.sh
hellt: @robshakir it doesn't direclty map to the protoc version, as there are more than just `protoc` in the package (namely grpc plugins for both py and go version).
Therefore, the idea is that `ghcr.io/srl-labs/protoc:0.0.3` is tagged in a semver fashion and the image tag matches the tag in the git repo for which the Dockerfile reveals the used versions: https://github.com/srl-labs/protoc-container/blob/v0.0.3/Dockerfile

robshakir: ACK - no problem. I was just wondering whether it is possible to directly map these together so that it's clearer, but it's written into the `pb.go` as to what versions are used so I don't see this is a huge issue.

The only downside here is that we stay pinned to that `0.0.3` version until an explicit change is made. If there's a semver convention here - could we consider tagging something like `1.x.x-latest`? (It'd be nice to stay close to HEAD for the generated files.)

n0shut: @robshakir good call, I see that the upstream changed the logic to include protoc__protoc-gen-go version in the image tag, so that it clear which sw components versions are used.

https://github.com/srl-labs/protoc-container#versioning

I have therefore update the branch in [4a11050](https://github.com/openconfig/gnmi/pull/136/commits/4a11050420ed9f7bd5a7f970134f2a2ccadfb855) to reflect that.

Now it is explicit which version of protoc and protoc-gen-go are used in the image.

compile-protos-docker.sh
```diff
@@ -0,0 +1,65 @@
...
+${DOCKER_CMD} \
+  bash -c "${PROTOC_GO_CMD} proto/collector/collector.proto"
+
+# generating proto/gnmi_ext/gnmi_ext.proto
+${DOCKER_CMD} \
+  bash -c "${PROTOC_GO_CMD_NOGRPC} proto/gnmi_ext/gnmi_ext.proto"
+
+# generating proto/target/target.proto
+${DOCKER_CMD} \
+  bash -c "${PROTOC_GO_CMD_NOGRPC} proto/target/target.proto"
```
n0shut, line 39: One thing I am not particularly happy with is that this script generated bindings for all the listed protos, which might be not desirable I think? For example, if only gnmi_ext has changed and you want to regenerate its bindings it would be cool to have a command that does just that.

What I normally do myself is writing a Make file with separate targets, something like this:

* `make gen-all` - generate bindings for all listed protos
* `make gen-fake.proto` - generate bindings for fake.proto
* etc

But since I do not see a Makefile in this repo I didn't want to introduce right away.

If make is not to ones liking, it can be a bash with functions, simulating the same behavior.

compile-protos-docker.sh
n0shut: I can also port it to gnoi repo where it is even more of a pressing issue.

n0shut:
Hi @robshakir 
I have updated this PR by introducing `./run.sh` script that embeds purpose-built functions for each lang/proto.
I have updated the initial PR comment to reflect that, but in essense what you can now do is:

```bash
# to compile all protos
./run.sh compile-all

# to compile selected protos
./run.sh compile-go-gnmi-ext
```

hellt:
tsk tsk tsk
still using this user unfriendly compile_protos.sh that has no reproducibility and heavily dependent on hosts' configuration

robshakir:
Thanks for this -- let me look again this week. Unfortunately, I was on an extended leave during the period you previously requested review.

hellt:
No worries @robshakir, I didn't mean to push, just a slight nudge. I can update it with the updated versions of protoc and go plugins

run.sh
```diff
@@ -0,0 +1,122 @@
...
+# to generate all protos: ./run.sh compile-all
+# to generate a specific proto for Go: ./run.sh compile-go-gnmi
+# to generate a specific proto for Python: ./run.sh compile-py-gnmi
+
+
+set -o errexit
+set -o pipefail
+
+# container image tag, to see all available tags go to
+# https://github.com/srl-labs/protoc-container/pkgs/container/protoc
```
robshakir, line 14: Is this container something that you expect to continue to be updated?

run.sh
```diff
@@ -0,0 +1,122 @@
...
+  wait
+  echo
+  echo "All compilations finished!"
+}
+
+# generating proto/gnmi/gnmi.proto
+function compile-go-gnmi {
+    ${DOCKER_CMD} ash -c "${PROTOC_GO_CMD} proto/gnmi/gnmi.proto"
+    echo "finished Go compilation for gnmi.proto"
+
```
robshakir, line 60: (nit) is this newline needed (here and below)?

robshakir (commented):
I'd propose to also leave `compile_protos.sh` present as well -- since there are some environments where security restrictions mean that docker cannot be run (e.g., locations that do not allow a virtual machine to be run).

robshakir:
@marcushines - PTAL too.

run.sh
n0shut: yes, I just bumped it to a more recent version tuple [227be07](https://github.com/openconfig/gnmi/pull/136/commits/227be073d1d6105d6e67548e74317bd7617a39a8)

run.sh
n0shut: fixed, thanks

n0shut:
compile_protos.sh brought back, @robshakir

n0shut:
closing as stale

PR #137, opened 2023-02-18
Example: Updated protos with container-based proto generator

n0shut:
This is an example of a product of compilation based on #136.

Note, how less drastic the changes are, compared to #135, while using the more recent version of `protoc`.
Avoiding the environment gravitation is the reason for #136. We likely want everyone to produce a repeatable and consistent generation product.


bstoll:
Closing as I believe this is stale now - the situation is a bit better now that we have a build toolchain pinned with Bazel.

PR #154, opened 2023-10-11
commit confirmed

hellt:
This PR adds a proposal for commit confirmed extension.

Extension description is provided in https://github.com/openconfig/reference/pull/198


proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -89,3 +92,47 @@ message TimeRange {
...
+// The CommitID is used in the subsequent SetRequest/CommitConfirm messages
+// to identify commit to be confirmed, reset or rejected.
+message CommitID {
+  string id = 1;
+}
+
+// The CommitConfirmed allows automatic rollback of a commit
+// if Set changes are not confirmed within the specified timeout.
+// The document about gNMI commit confirmed can be found at
+// https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-commit-confirmed.md
```
ncorran, line 108: Is there a working version of this file?

hellt: yes, it is a PR against `reference` repo - https://github.com/openconfig/reference/pull/198

proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -89,3 +92,47 @@ message TimeRange {
...
   int64 end = 2;   // Nanoseconds since the epoch
 }
+
+// The CommitID is returned by the gNMI server to indicate the ID of a commit.
+// It is returned in response to the Set Request with the CommitConfirm message
+// that had COMMIT_CONFIRMED_ACTION_START set.
+// The CommitID is used in the subsequent SetRequest/CommitConfirm messages
+// to identify commit to be confirmed, reset or rejected.
+message CommitID {
+  string id = 1;
```
ncorran, line 102: What is the level of security on the commit id here?
- Can I accidentally or maliciously confirm or reject someone else's commit?
What is the scope
- Is the commit id expected to be valid only within some scope?
What checks can a server make to see that the CONFIRM/REJECT request comes from the same client as the START (other than commit id match); can we assume the following request will come from a connection/session with the same parameters?
Can any number of commit confirms be open at once? (What is a server wanted to set a max limit)?
Should there be a max timeout, or could a server impose one?  (If it can, should the  CommitID message also include a duration in the response so the client knows what time they have?  Similarly would it be useful to return the server default duration for the case where 0 is input?)

hellt: 1. a user executing gNMI RPC is authenticated using the [standard auth mechanisms](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-authentication.md). It might depend on the implementation, but one can expect that a commit confirmed started by user U1 can only be confirmed by U1 user.
2. I am not aware of NOSes that support multiple parallel commits. If there are NOSes that support multiple parallel commits this spec must be adjusted.

proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -30,6 +32,7 @@ message Extension {
     // Well known extensions.
     MasterArbitration master_arbitration = 2;  // Master arbitration extension.
     History history = 3;                       // History extension.
+    CommitConfirmed commit_confirmed = 4;      // Commit Confirmed extension.
```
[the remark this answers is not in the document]
hellt: This is common to any extension.
Rejects and Confirms are sent within the SetRequest message.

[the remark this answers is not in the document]
ncorran: Thanks - is also clear in the linked .md you referenced

proto/gnmi_ext/gnmi_ext.proto
hellt: The proposed commit-confirmed workflow assumes that once the commit-confirmed started, the lock is put in place preventing any other commits from happening. 

Any subsequent SetRequest without commit-confirmed extension (a regular SetRequest) will yield an error. This is not explained in https://github.com/openconfig/reference/pull/198 and you made a great point that it needs to be clarified.

Because no parallel commit-confirmed are allowed, it is implicitly assumed that the running datastore is locked for any other commits until the commit in question is confirmed or rejected

proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -89,3 +92,47 @@ message TimeRange {
...
+message CommitID {
+  string id = 1;
+}
+
+// The CommitConfirmed allows automatic rollback of a commit
+// if Set changes are not confirmed within the specified timeout.
+// The document about gNMI commit confirmed can be found at
+// https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-commit-confirmed.md
+message CommitConfirmed {
+  CommitConfirmedAction action = 1;
```
kidiyoor, line 110: What is the rational behind defining _action_ as an enum over what is proposed in https://github.com/openconfig/reference/issues/197 which is to use `oneof id` which implicitly models the 3 actions - commit/confirm/cancel(START/CONFIRM/REJECT) and also implicitly enforces required behavior at the proto level, mandating the request to be only one of these actions at a time.

proto/gnmi_ext/gnmi_ext.proto
ncorran: Great - yes please clarify in the document, but also I think it would be better to make this explict here with the action names or at least the comments next to the action names:
START has to also LOCK
REJECT/CONFIRM have to UNLOCK
TIMEOUT has to be handled as REJECT (rollback config and also UNLOCK)

hellt: I will defer for now and wait for OC/Google comments on this PR since it can be they would like to proceed with their version which is in https://github.com/openconfig/gnmi/pull/155

dplore:
> @dplore does it make sense to ask for the community to cast votes/comment for #154 and #155 to move the needle forward?

We can take comments now and also add to the agenda for next OC Community review (first Thursday of each month).  

The decision to merge will be made by the OC Operators group which has weekly private meetings.  We will review and compare both at the Oct 24th occurrence of the OC Operators meeting.

proto/gnmi_ext/gnmi_ext.proto
kidiyoor: To eliminate confusion of the spec appearing to use different IDs - I have updated the `oneof` to represent `actions` instead of IDs - https://github.com/openconfig/gnmi/pull/155/files. With this change both the proposal looks identical in terms of semantics and functionality.

I believe that using `oneof` over `ENUM` for this particular spec helps define actions much more clearly. Eg. `rollback_duration` is only valid during commit. `oneof` enforces that behavior, with ENUMs client can pass `rollback_duration` field during confirm or cancel request. Its best to avoid unnecessary explanation in documentation  if we can enforce them in the spec which makes the API much cleaner.

proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -89,3 +92,47 @@ message TimeRange {
...
+enum CommitConfirmedAction {
+  // gNMI server should return InvalidArgument error if the action is not
+  // specified for a CommitConfirmed extension.
+  COMMIT_CONFIRMED_ACTION_UNSPECIFIED = 0;
+  // Starts a commit with a specified rollback timeout
+  // or resets the timeout for the commit that has confirmation pending.
+  COMMIT_CONFIRMED_ACTION_START = 1;
+  // Confirm a commit identified by CommitID.
+  COMMIT_CONFIRMED_ACTION_CONFIRM = 2;
+  // Reject a commit identified by CommitID
```
dplore, line 133:
```suggestion
  // Reject and rollback a commit identified by CommitID
```

dplore:
OC Operators reviewed today.  There is some slight preference for #154 based on slightly better/more explicit documentation for each field in the proto.  The protos seem equivalent in functionality.   

An open question for the community is, is #154 is missing anything?   Please provide comments in the next two weeks, we'll move to merge this on Nov 7, 2023

kidiyoor:
It looks like there are two key differences b/w #154 and #155, we should probably clarify for posterity why we are choosing one over the other.

1. Identifies(ID) generation

As per #154, The Commit ID is generated by the server. Client receives the ID in the response and uses this Commit ID to confirm/cancel the request. 

As per #155, that server should not be involved in ID management. Clients should generate and store the ID locally before making the SetRequest, this ID is passed during the request and server registers it, the same ID is used by the clients to confirm/cancel the request.

Strong opinion: There are already systems which deal with configuration generation and have an identifier associated with the generated configuration. We should allow such system to use the same identifier as the commit id. This will enhance the debug ability and visibility of the configuration flow throughout the pipeline.

2. Use of ENUM vs oneof

I am ok with either, as long as state the reason for the choice.

Opinion: It looks to me that `oneof` is able to model the three action in a more readable fashion without requiring much explanation. Specially if you look at field like `rollback_duration` which is only applicable during request, we are able to ensure it in the spec itself. However, this is not a strong opinion, if the community thinks its ok to use ENUM here that's fine.

proto/gnmi_ext/gnmi_ext.proto
hellt: @kidiyoor @dplore I think one of the fundamental differences between #154 and #155 is that in our proposal, the `CommitID` message is returned by the gNMI server. In other words, the server assigns Commit IDs and not the client selecting them.

This matched our implementation on both platforms (SR Linux and SR OS), and I feel like this is also true for most other NOSes that currently support commit operations.

hellt:
replaced by #155

PR #155, opened 2023-10-11
Commit confirmed extension proto

kidiyoor:
This is a supporting PR for https://github.com/openconfig/reference/issues/197.  Older discussion and reference/handling document PR can be found here: https://github.com/openconfig/reference/pull/196

Netconf [rfc6241](https://datatracker.ietf.org/doc/html/rfc6241#section-8.4.1) has a definition for `confirmed commit`. as follow-
```
  If the <persist> element is not given in the confirmed commit
   operation, any follow-up commit and the confirming commit MUST be
   issued on the same session that issued the confirmed commit.  If the
   <persist> element is given in the confirmed <commit> operation, a
   follow-up commit and the confirming commit can be given on any
   session, and they MUST include a <persist-id> element with a value
   equal to the given value of the <persist> element.
```

Example implementations of confirmed commit(There are other vendors who support confirmed commit but not listed here)-
1. Cisco - https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r3-9/system_management/command/reference/yr39xr12k_chapter5.html#wp1873946320 - Confirmed commit can be issued as part of CLI cmds - ` commit confirmed <seconds>` eg. `RP/0/0/CPU0:router(config)# commit confirmed 60`.  The confirmation can be issued by entering a `commit` in the same SSH session.

2. Juniper - https://www.juniper.net/documentation/us/en/software/junos/cli/topics/ref/command/commit.html - Confirmed commit can be issued as part of CLI cmds - `commit confirmed <minutes>`. Eg. `user@host# commit confirmed 1`. The confirmation can be issued by entering a `commit` or `commit check` in the same SSH session.

There is an _optional_ `comment` field that can be passed  which can hold any client provided description.

### DIFF
--- a/proto/gnmi_ext/gnmi_ext.proto
+++ b/proto/gnmi_ext/gnmi_ext.proto
@@ -21,6 +21,8 @@ syntax = "proto3";
 // extensions defined outside of this package.
 package gnmi_ext;
 
+import "google/protobuf/duration.proto";
+
 option go_package = "github.com/openconfig/gnmi/proto/gnmi_ext";
 
 // The Extension message contains a single gNMI extension.
@@ -30,6 +32,7 @@ message Extension {
     // Well known extensions.
     MasterArbitration master_arbitration = 2;  // Master arbitration extension.
     History history = 3;                       // History extension.
+    Commit commit = 4;                         // Commit confirmed extension.
   }
 }
 
@@ -88,4 +91,24 @@ message History {
 message TimeRange {
   int64 start = 1; // Nanoseconds since the epoch
   int64 end = 2;   // Nanoseconds since the epoch
+}
+
+// Commit confirmed extension allows automated revert of a configuration after
+// certain duration if an explicit confirmation is not issued. This duration can
+// be used for testing the network device to verify the configuration.
+// It also allows explicit cancellation of the commit during the rollback duration
+// time window if client determine that the configuration needs to be reverted.
+// The document about gNMI commit confirmed can be found at
+// https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-commit-confirmed.md
+message Commit {
+  // Maximum duration to wait for a confirmaton before reverting the commit.
+  google.protobuf.Duration rollback_duration = 1;
+  oneof id {
+    // commit_id indicates that this is a new commit request with given ID.
+    string commit_id = 2;
+    // confirm_id indicates that this is a confirmaton request for the given ID.
+    string confirm_id = 3;
+    // cancel_id indicates that this is a cancellation request for the given ID.
+    string cancel_id = 4;
+  }
 }
--- a/proto/gnmi_ext/gnmi_ext.proto
+++ b/proto/gnmi_ext/gnmi_ext.proto
@@ -108,9 +108,10 @@ message Commit {
   }
 }
 
-// Create a new commit request.
+// Create a new commit confirmed request.
 message CommitRequest {
-  // Commit id for the request.
+  // Commit id for the request. ID must be generated at the client side, it can be used
+  // to confirm or cancel the request without the rollback_duration interval.
   string id = 1;
   // Maximum duration to wait for a confirmaton before reverting the commit.
   google.protobuf.Duration rollback_duration = 2;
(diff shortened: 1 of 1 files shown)


proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -89,3 +92,38 @@ message TimeRange {
...
+    CommitRequest commit = 1;
+    CommitConfirm confirm = 2;
+    CommitCancel cancel = 3;
+  }
+}
+
+// Create a new commit request.
+message CommitRequest {
+  // Commit id for the request.
+  string id = 1;
```
hellt, line 114: @kidiyoor what are the semantics behind this `id`?
Is it returned by the server when CommitRequest is issued?

kidiyoor: It is passed by the client making the commit request. If the id is not passed its a bad request.

proto/gnmi_ext/gnmi_ext.proto
```diff
@@ -89,3 +92,42 @@ message TimeRange {
...
+    // commit action creates a new commit. If a commit is on-going, server returns error.
+    CommitRequest commit = 2;
+    // confirm action will confirm an on-going commit, the ID provided during confirm
+    // should match the on-going commit ID.
+    CommitConfirm confirm = 3;
+    // cancel action will cancel an on-going commit, the ID provided during cancel
+    // should match the on-going commit ID.
+    CommitCancel cancel = 4;
+  }
+}
```
hellt, line 118: @kidiyoor, to pick up on our discussion in OC community call: We at Nokia can support this mode of operation where a client chooses the ID for he commit confirm operation.

I wonder, though, if we should consider adding `CommitResponse` message that clients can provide in the SetResponse extension field to allow servers to respond with custom data that clients MAY be interested in.
This data may be `bytes` and can contain something like internal commit ID that we discussed.

Again, this `CommitResponse` message is optional, servers might choose not to set it up at all.

WDYT?

kidiyoor: I would be careful with introducing a field which don't have a standard behavior across vendors in the spec. Since the goal of gnmi is to standardize the API across vendors, with such an optional response there is possibility of client/server implementation which is not interoperable.

If we can define a common use-case, with dedicated fields(maybe not bytes), then I think we can definitely consider.

kidiyoor: Can you talk more about the use-case here (is it same as the subscription use-case that was mentioned during community call)?

If the data expected form the server is going to be server generated ID - then is it only relevant to "confirmed commit" use-case ? should the response data be part of regular Set commit as well ? can you talk about why its only specific to "confirmed commit" ?

hellt: @kidiyoor I was just thinking that returning back internal ID might be useful down the road.
But I am ok dropping this and us (nokia) using a custom extension instead.

dplore: @hellt if you can describe the `CommitResponse` message in proto, then I think this could be modeled. I think this would make sense to in a follow up PR, unless you think it would conflict with this PR as is.

hellt: @dplore yes, we can follow up in the next PR

dplore:
Please add 2 references to documentation of similar capability and add to the first comment in this PR.

dplore (approved):
This was reviewed in the 14-Nov-2023 OC Operators meeting without objection.  Issuing a last call for two weeks (ending 28-Nov-2023) for any other comments on this change.

hellt:
It is a pity we haven't addressed the timer reset @dplore

dplore:
Hi @hellt please submit a new PR to add the timer reset.  Be sure to include references.  It seems like a straight forward addition?  We can wait for the timer reset to be merged before creating a new gnmi release tag.

proto/gnmi_ext/gnmi_ext.proto
Ashu008: can i use gnmic tool to request for gnmi commit ?

PR #165, opened 2024-01-10
TechDebitReduction: Cleanup some of the language in the documentation.

morrowc:
Make things consistent (use of ns/nanos/nanoseconds, use of modules/models). Remove some deprecated options expressed in the documentation.

### DIFF
--- a/proto/gnmi/gnmi.proto
+++ b/proto/gnmi/gnmi.proto
@@ -20,8 +20,8 @@ import "google/protobuf/descriptor.proto";
 import "github.com/openconfig/gnmi/proto/gnmi_ext/gnmi_ext.proto";
 
 // Package gNMI defines a service specification for the gRPC Network Management
-// Interface. This interface is defined to be a standard interface via which
-// a network management system ("client") can subscribe to state values,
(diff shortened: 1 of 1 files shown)


proto/gnmi/gnmi.proto
```diff
@@ -206,8 +206,8 @@ message Decimal64 {
 // ScalarArray is used to encode a mixed-type array of values.
 message ScalarArray {
   // The set of elements within the array. Each TypedValue message should
-  // specify only elements that have a field identifier of 1-7 (i.e., the
-  // values are scalar values).
+  // specify only elements that have a field identifier of 1-5 (i.e., the
+  // values are not deprecated scalar values).
```
hellt, line 210: this seems to be (1-5,14), where 14 is a double_val

but maybe it is not necessary to call it out in the comment at all, and just say that "...only non deprecated scalar values"

morrowc: like that simplification,thanks!

morrowc:
Oddly I can't assign a code-owner for review... perhaps I need to do this all on the 'internal' version first ;(

PR #164, opened 2024-01-09
Added CommitSetRollbackDuration action for Commit Confirmed extension

hellt:
This is an alternative version to #161.
In this version, a specific action is added to denote the Set Rollback Duration for the Commit Confirmed action.

PR #188, opened 2024-10-09
Project import generated by Copybara

balajisundaravel:


proto/target/target.pb.go
```diff
@@ -16,8 +16,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
-// 	protoc        v4.24.0
+// 	protoc-gen-go v1.33.0
+// 	protoc        v3.21.12
```
dplore, line 20: Does it matter that an older version of protoc was used to generate the protobufs?

balajisundaravel: yeah. I tried to update the protobuf compiler and it seems that `3.21.12` is the latest version.
 ```
 protobuf-compiler is already the newest version (3.21.12-8.2).
Summary:
  Upgrading: 0, Installing: 0, Removing: 0, Not Upgrading: 0
```

Unless I am missing something there.

dplore (approved):
LGTM,  @robshakir can you take a second look?

In a future PR, I'll import https://github.com/openconfig/gnmi/pull/164 into the upstream repository and re-update this repo.   But #164 doesn't have any conflict with the update (other than the generated gnmi_ext which are currently  out of date are still out of date)

balajisundaravel:
@dplore thank you for merging the changes.

PR #189, opened 2024-11-13
bazel: Migrate to bzlmod for dependency management

minor-fixes:
This change aims to fix the `bazel build` postsubmit workflow in a somewhat heavy-handed way: by migrating to use bzlmod to manage dependencies, and by updating said dependencies (and the supported version of bazel itself) to the newest version(s) feasible.

This involves:

* Removing the old dependency fetching infra from `WORKSPACE.bazel`, and adding equivalent dependency fetches to `MODULE.bazel`. Each version used is the newest available on [BCR](https://registry.bazel.build), with the exception of:
  - `rules_proto`, which encounters a Starlark error with the latest version of `7.0.2`
  - `protobuf`, which must match versions with the toolchain loaded by `toolchains_protoc`; the latest version of `toolchains_protoc` does not have a Starlark mapping for protobuf v29.0 yet.
* Changing rules used to generate C++ protobuf/gRPC bindings from `cc_grpc_library` (provided by gRPC) to `cpp_grpc_library` (provided by `rules_proto_grpc_cpp`), which seems to be the [current canonical ruleset](https://bazel-contrib.github.io/SIG-rules-authors/proto-grpc.html#the-ruleset-options) for this. These rules generate proto, grpc, and mock bindings by default, so the corresponding attributes are dropped since they're no longer necessary.
* Dropping `gnmi_deps.bzl`, as it is no longer used to fetch dependencies, and can only confuse downstream users that also load this file.

Additionally, there are some quality-of-life changes:

* `.bazelversion` is added to pin the version of bazel used when invoked via `bazelisk`
* `.gitignore` is added to avoid committing bazel-generated symlinks
* Go protobuf generation is added for all protos

Tested: `bazel build //...` works on my system (Linux, amd64)

---

There are good reasons not to take this change, including:

* C++ proto rule changes causing downstream breakage, as some targets are removed, and the semantics of remaining targets may have changed in an incompatible way
* Downstream bazel repositories will need to migrate to bzlmod (at least partially) in order to pick up updates following this change

...and this change author has little context on the presence/severity of these issues - guidance here would be welcome!


minor-fixes:
startblock:
  add reviewer @marcushines ;)

bstoll:
I managed to miss this PR but I just created #192 which encompasses most of these changes you are proposing here.

With the recent changes to deprecate rules_proto and move Bazel rules into the GRPC/Protobuf packages this whole situation is quite confusing. My proposed change keeps WORKSPACE support and C++ targets the same - I am not even sure if this was possible at the time you wrote your PR to maintain backwards compatibly and support bzlmod.

Issue #95, opened 2021-04-24
gnmi_ext proto error when doing protoreflect 


sbapu-arista:
We are seeing issues with the way gnmi_ext.pb.go is generated from gnmi_ext.proto
When using https://github.com/jhump/protoreflect for doing reflection on gnmi, we are getting this error

File not found: github.com/openconfig/gnmi/proto/gnmi_ext/gnmi_ext.proto

Digging a little more we found that the file descriptor generated in gnmi_ext.pb.go
here at line 549
549 var File_proto_gnmi_ext_gnmi_ext_proto protoreflect.FileDescriptor
Has a relative path
21  // source: proto/gnmi_ext/gnmi_ext.proto
as indicated in line 21 of the file
While the golang import path is 
github.com/openconfig/gnmi/proto/gnmi_ext/
as imported in github.com/openconfig/gnmi/proto/gnmi/gnmi.pb.go

It would be great if we can generate gnmi_ext.pb.go with relative path 
github.com/openconfig/gnmi/proto/gnmi_ext/gnmi_ext.proto
instead of 
proto/gnmi_ext/gnmi_ext.proto

We were able to test that the issue gets resolved by regenerating gnmi_ext.pb.go with the said relative path.

jxx-gg:
While looking, just curious what kind of file description information are you looking at? Will the existing one implemented by default (google.golang.org/protobuf/reflect/protoreflect) meet your requirement?  e.g. 
```
e := gnmi_ext.Extension{}
d := e.ProtoReflect().Descriptor() 
```

sbapu-arista:
Hi Johnson,
This is not exactly a problem with opeconfig/gnmi but with the actual protoreflect tools. Google's protoreflect being one of them. I'm hoping this issue will make it clear
https://github.com/golang/protobuf/issues/567

A better explanation of what we are facing
We use gRPC reflection to get the proto from services(when calls are made over plain http e.g: restful wrappers for grpc). As part of resolving a service, protoreflect resolves all of the imports in the file that the service is contained in[1]. However, this resolution is exact match, so the name of the imported file must match what the path the proto file descriptor is registered[2] as. The path the proto-file descriptor is registered as is path to the file passed to protoc.

[...]

PR #192, opened 2025-01-14
Update build to support Bazel 8/bzlmod

bstoll:
This PR updates GNMI proto generation to be more closely tied to Bazel, similar to other repos like GNOI with the regenerate-files.sh script.

  * Go toolchain is now pinned with Bazel.
  * bazel/tools.go is added to pin go proto compilers into go.mod
  * BUILD files are added for all packages (generated by Gazelle).
  * Python proto support is removed.
  * GRPC/Proto toolchains updated to latest releases.
  * protos use import_prefix - this should help these issues when using reflection and paths are relative to each repo instead of globally unique: openconfig/gnoi#216 , openconfig/gnsi#176 , openconfig/gribi#37 , openconfig/gnmi#95
  * Gazelle target at root will auto generate BUILD files for Go.

This PR is the Copybara output in Google from cl/707303444.


marcushines (approved):

vincentbernat:
There is an unexpected bump to the required Go version in `go.mod`.

PR #203, opened 2025-07-11
Add Renovate configuration file.

bstoll:
Add configuration for [renovate](https://www.mend.io/renovate/). The intent is to use renovate instead of dependabot for helping with dependency chores. Renovate supports updating bazel/go toolchains, go modules, bzlmod, docker, and Github Actions, while dependabot only supports updating go modules.

We will use [Forking Renovate](https://github.com/apps/forking-renovate) so we do not need to give direct write permissions to the bot.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,20 @@
+{
+  "extends": ["config:recommended", "schedule:weekdays", ":semanticCommitsDisabled"],
+  "postUpdateOptions": ["gomodTidy"],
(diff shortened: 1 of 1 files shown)


dplore (approved):
The idea sounds good to me.  I don't understand how the json here causes renovate to be used?  Is there some setting in the repository that is updated to move from dependabot to renovate?

bstoll:
Closing; this change was done in #204.

PR #224, opened 2025-09-29
Commence deprecation of suppress_redundant

earies:
Following up on the long standing PR: https://github.com/openconfig/gnmi/pull/114

This PR is to request for comment and implementation status as to the overall
deprecation of `suppress_redundant` within gNMI subscriptions.

Should this change be accepted, the gNMI specification will be updated as well
as relevant .proto and generated stubs.


ncorran:
Happy to see this be removed.  Cisco-XR dont support the option (and don't have any plans to support it in future)

jsterne:
Happy to see this removed on the Nokia side.

dplore:
> @dplore
> 
> > As far as deprecating existing gNMI specification, we should understand is what dependencies or use cases exist regarding suppress_redundant and heartbeat_interval. It would be good to hear from the community if these are commonly implemented and in use features.
> 
> We at IP Infusion have suppress_redundant feature ready for shipping. Yes, implementation of this is involved, but worth it since the feature is very useful as it reduces the amount of data that is pushed to the collector significantly in most use-cases.
> 
> @earies I did not see any good reason in this thread that warrants deleting this feature from the gnmi spec.

Hi @raghubk , thank you for your comment.  I did a quick search of  IP Infusion documentation and I don't see any reference to configure a suppression of redundant values for telemetry.  Can you show us evidence that this is a supported feature?

(I looked here in your release SP 6.6 for [streaming telemetry commands](https://documentation.ipinfusion.com/ocnos-streaming-telemetry-6.6/Content/Streaming-telemetry/telemetry-commands.htm?tocpath=Streaming%20Telemetry%7CStreaming%20Telemetry%20Commands%7C_____0
))

dplore:
> If we deprecate suppress redundant, how do you plan to implement the 'dynamic value' paths like counters, temp, cpu, ram and similar sensors. If its just sampling, its not optimal as we can reduce by using suppress redundant as there can be long periods that these sensors remain same and then change. If its on-change, then we are actually not doing on change, but suppress redundant internally with a small interval, as its practically not possible to monitor these at very fine granularity without impacting the system performance.

Hi @santanukar2000,  we are asking for evidence that this feature is implemented and even more importantly, in use in production networks.  I agree that the feature is theoretically useful.  But if there are no implementations and therefore no operational usage after so many years of being present in the gnmi specification, it is suggested by the community that it be deprecated.

earies:
Hi folks - it has been some months since there has been any activity on this PR and have not seen any warranted operational or implementation evidence.  I'll suggest commencing a final call and if necessary a quick agenda item for Apr 2026 community call if we need further discussion.

dplore:
@raghubk and @santanukar2000 please share any documentation you have about ipinfusion support for this feature for the community benefit.

raghubk:
@dplore Currently the customer release of this feature is delayed due to other factors. We think end of year they will evaluate streaming telemetry.

eschornig-tech:
This functionality is supported by NVIDIA switches for a number of releases. See here: https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-516/Monitoring-and-Troubleshooting/gNMI-Streaming/

ElodinLaarz:
Discussed at OC Community Meeting May 7th, 2026:

* The reason for deprecating the flag is because vendors (on the whole) do not plan to ever support it.
* Since it's part of the gNMI contract, then there's an expectation that vendors will support all base features.

Routers would need to keep state, which has proven to be infeasible for many vendors. The advantage of using suppress redundant in many cases is not necessarily outweighed by the added memory footprint.

At least 2 vendors (from the above comments) have successfully implemented it, which for OpenConfig / gNMI is enough to be allowed in the model. 
An alternate choice here is to notate the suppress_redundant field can marked as 'genuinely optional' in the proto.

Deprecation does not mean that the leaf cannot be used, but going forwards, adding implementations would be discouraged.

dplore:
Thanks for all the thoughtful discussion on this at the [May 7, 2026 OpenConfig community meeting](https://docs.google.com/document/d/11sXUF2zujhbPs_bt-RFJhh2DcIbEvnOpd9GypE-u3cc/edit?tab=t.0#heading=h.k2658yxgiun5).

Given there are at least 2 implementations of `suppress_redundant` which include customer usage I think we should not formally deprecate the field.  

I suggest the textual description for `suppress_redundant` be updated to indicate that this field is optional.  The intent is to indicate that an implementation can be fully compliant with the gnmi proto and not implement this field.

